PR #105, opened 2026-03-06
⚡ Bolt: Optimize DiffViewer rendering to prevent cascading re-renders

madara88645:
💡 **What**: Refactored `DiffViewer` component to use `useMemo` instead of `useState` and `useEffect` for the text diffing computation.
🎯 **Why**: Previously, `DiffViewer` computed the text diffs inside a `useEffect` hook, which then triggered a state update (`setHtmlContent`). This pattern causes unnecessary cascading re-renders (an initial empty render, followed by the effect, then a second render with content).
📊 **Impact**: Saves one complete render cycle for the component every time its props change. It also correctly caches the expensive string comparison logic (`dmp.diff_main`) in case of unrelated parent re-renders. It also resolves a React ESLint warning (`react-hooks/set-state-in-effect`).
🔬 **Measurement**: Component renders without a cascading second pass, and linting passes without the `set-state-in-effect` error. Verified visually via Playwright test that the diff view still renders perfectly.

---
*PR created automatically by Jules for task [5012538211463277399](https://jules.google.com/task/5012538211463277399) started by @madara88645*

### DIFF
--- a/web/app/components/DiffViewer.tsx
+++ b/web/app/components/DiffViewer.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useMemo } from "react";
+import { useDeferredValue, useMemo } from "react";
 import { diff_match_patch, DIFF_INSERT, DIFF_DELETE, DIFF_EQUAL } from "diff-match-patch";
 
 interface DiffViewerProps {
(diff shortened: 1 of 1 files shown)


madara88645:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/madara88645/Compiler/pull/105#pullrequestreview-3902541487)

PR #217, opened 2026-03-21
CI-Control Merge: security fixes, accessibility, optimizations, and t…

madara88645:
…est infrastructure

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -26,6 +26,7 @@ from app.emitters import (
     emit_user_prompt_v2,
     emit_plan_v2,
     emit_expanded_prompt_v2,
+    _is_trivial_input,
 )
 from app.rag.simple_index import (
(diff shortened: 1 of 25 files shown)


madara88645:
Embedding an API key via NEXT_PUBLIC_API_KEY and adding it to every browser request (x-api-key) exposes the key to all users (it’s bundled into the client build) and enables trivial replay outside the app. If this key is meant to protect privileged endpoints, it must be kept server-side (e.g., API_KEY without NEXT_PUBLIC_), and the client should call a Next.js route handler / backend proxy that injects the secret header. If the endpoints are intended to be public, consider removing the API-key requirement instead of shipping a shared key to the browser.
web/app/components/DiffViewer.tsx:1 this must be removed @openai-code-agent

PR #325, opened 2026-04-03
🎨 Palette: Add keyboard shortcuts to Generator pages

madara88645:
💡 **What:** 
Added `Ctrl/Cmd+Enter` keyboard shortcuts to submit prompts directly from the textareas in the Agent Generator and Skills Generator pages. Added visual `<kbd>` hints to the generate buttons, alongside proper `focus-visible` states.

🎯 **Why:** 
To allow power users to remain on the keyboard during the generation workflow without needing to switch to the mouse to click "Generate". The visual `<kbd>` hints make this feature discoverable, and the `focus-visible` styling improves navigation for pure keyboard users.

📸 **Before/After:**
*(See attached video verification in review)*
- **Before:** Textareas could only grow; submit required manual mouse clicking or awkward tabbing. Buttons lacked clear focus rings.
- **After:** Hitting `Ctrl/Cmd+Enter` immediately triggers generation. Submit buttons now display "Ctrl/⌘ Enter" styled cleanly on the right, and tab-focus displays a crisp offset ring.

♿ **Accessibility:**
- Added `focus-visible:ring-2`, `focus-visible:outline-none`, and `focus-visible:ring-offset-2` to the submit buttons for explicit keyboard focus indication.
- Used `<kbd>` semantic markup for the keyboard shortcuts.

---
*PR created automatically by Jules for task [17506313857645639799](https://jules.google.com/task/17506313857645639799) started by @madara88645*

### DIFF
--- a/web/app/agent-generator/page.tsx
+++ b/web/app/agent-generator/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState } from "react";
+import { useRef, useState } from "react";
 import ReactMarkdown from "react-markdown";
 import { apiFetch, buildGeneratorApiHeaders } from "@/config";
 import ContextManager from "../components/ContextManager";
@@ -16,8 +16,12 @@ export default function AgentGenerator() {
   const [includeExampleCode, setIncludeExampleCode] = useState(false);
   const [history, setHistory] = useState<{ label: string; prompt: string }[]>([]);
 
+  const isGeneratingRef = useRef(false);
+
   const handleGenerate = async () => {
     if (!description.trim()) return;
+    if (isGeneratingRef.current) return;
+    isGeneratingRef.current = true;
 
     setLoading(true);
     setError(null);
@@ -51,6 +55,7 @@ export default function AgentGenerator() {
       setError(e instanceof Error ? e.message : "Failed to generate agent");
     } finally {
       setLoading(false);
+      isGeneratingRef.current = false;
     }
   };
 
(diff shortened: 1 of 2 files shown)


madara88645 (commented):
**Not ready to merge — concurrency bug in keyboard shortcut handler.**

Copilot flagged a real issue: the `Ctrl/Cmd+Enter` keydown handler can fire multiple concurrent API requests before React's state update flushes the `loading` flag to `true`.

**Root cause**: Reading `loading` from state inside the keydown handler is stale — `setLoading(true)` inside `handleGenerate` does not synchronously update the closure's copy of `loading`, so rapid keypresses or held-Enter auto-repeat can invoke `handleGenerate()` multiple times in the same render cycle.

**Suggested fix**: 
- Check `e.repeat` in the keydown handler to ignore auto-repeat events.
- Use a `useRef` in-flight guard (e.g., `isGeneratingRef.current`) that is set synchronously before the async call and cleared in the `finally` block — refs update immediately without a re-render cycle.

This affects both `web/app/agent-generator/page.tsx` and `web/app/skills-generator/page.tsx`.

madara88645:
@copilot Please fix the concurrency bug in both `web/app/agent-generator/page.tsx` and `web/app/skills-generator/page.tsx`:

The `Ctrl/Cmd+Enter` keydown handler can trigger multiple concurrent `handleGenerate()` calls because React state (`loading`) doesn't update synchronously between renders.

**Fix**:
1. Add `if (e.repeat) return;` at the top of the keydown handler to ignore auto-repeat when the key is held down.
2. Add a `useRef` in-flight guard so only one request runs at a time:
   ```tsx
   const isGeneratingRef = useRef(false);

   // Inside handleGenerate, at the very top:
   if (isGeneratingRef.current) return;
   isGeneratingRef.current = true;
   // ... existing logic ...
   // In the finally block:
   isGeneratingRef.current = false;
   ```

Apply both changes to both generator pages.

PR #326, opened 2026-04-04
[codex] Gate Fly.io deploy on successful CI

madara88645:
## What changed
- switched Fly.io deploy workflow from `push` on `main` to `workflow_run` after the `CI` workflow completes
- gated the deploy job so it only runs when the upstream CI conclusion is `success` and the source branch is `main`
- added a regression test that loads `.github/workflows/deploy.yml` and asserts the deploy workflow remains CI-gated

## Why
The deploy workflow was running independently of CI, so a broken `main` push could still reach production. This change makes deployment depend on a successful CI run and keeps the protection covered by a test.

## Impact
- production deploys now wait for CI success
- future workflow drift should be caught by the new regression test

## Validation
- `python -m pytest -q tests/test_deploy_workflow.py`
- `python -m pre_commit run --files .github/workflows/deploy.yml tests/test_deploy_workflow.py`

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,14 +1,16 @@
 name: Deploy to Fly.io
 
 on:
-  push:
-    branches: [ main ]
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
 
 jobs:
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
-    needs: []  # CI testlerini beklemek için: needs: [test] — ama ci.yml ayrı workflow, bu yüzden şimdilik bağımsız
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main' }}
     concurrency:
       group: deploy-production
       cancel-in-progress: true
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,13 +10,15 @@ jobs:
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
-    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main' }}
+    if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main' && github.event.workflow_run.event == 'push' }}
     concurrency:
       group: deploy-production
       cancel-in-progress: true
 
     steps:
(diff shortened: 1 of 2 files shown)


madara88645 (commented):
**Not ready to merge — two correctness issues with the `workflow_run` trigger.**

Copilot flagged issues that need to be addressed:

1. **Scheduled CI runs trigger deploys**: `workflow_run` fires for *all* CI completions, including scheduled runs (e.g., nightly). A successful scheduled CI run on `main` would deploy to production. Add `github.event.workflow_run.event == 'push'` to the `if:` condition to restrict deploys to push-triggered CI runs only:
   ```yaml
   if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main' && github.event.workflow_run.event == 'push' }}
   ```

2. **Deploy may use a different SHA than CI tested**: With a `workflow_run` trigger, `actions/checkout` uses the default ref of the deploy workflow — if new commits land on `main` between CI finishing and the deploy job starting, you may deploy unvalidated code. Pin the checkout to `github.event.workflow_run.head_sha`:
   ```yaml
   - uses: actions/checkout@v4
     with:
       ref: ${{ github.event.workflow_run.head_sha }}
   ```

3. **Minor**: `tests/test_deploy_workflow.py` constructs the workflow path relatively; use `Path(__file__).resolve().parents[1] / '.github/workflows/deploy.yml'` for robustness when pytest runs from a non-root directory.

madara88645:
@copilot Please fix the three issues in `.github/workflows/deploy.yml` and `tests/test_deploy_workflow.py`:

1. **Restrict `workflow_run` to push-triggered CI only** (prevents scheduled CI runs from deploying):
   ```yaml
   if: ${{ github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.head_branch == 'main' && github.event.workflow_run.event == 'push' }}
   ```

2. **Pin checkout to the exact SHA that CI tested** (prevents deploying commits that landed after CI finished):
   ```yaml
   - uses: actions/checkout@v4
     with:
       ref: ${{ github.event.workflow_run.head_sha }}
   ```

3. **Fix the test path** in `tests/test_deploy_workflow.py` to use an absolute path so pytest works from any directory:
   ```python
   from pathlib import Path
   workflow_path = Path(__file__).resolve().parents[1] / '.github' / 'workflows' / 'deploy.yml'
   ```
   Also update the regression test assertions to cover the new `event == 'push'` condition and the `ref: head_sha` checkout.

PR #365, opened 2026-04-11
[codex] Fix implied persona matching pipeline

madara88645:
﻿## Summary
- normalize implied persona keywords before compiling regex patterns, matching the existing lowercase input path
- prefer more specific implied persona matches when multiple patterns have the same confidence score
- allow implied personas to replace only generic/default roles in the real `compile_text_v2` pipeline
- add regression coverage for handler-level persona detection plus local `/compile` IRv2 output

## Root cause
`detect_implied_persona()` lowercases prompt text before matching, but `_compiled_implied_personas` previously preserved mixed-case keys from `IMPLIED_PERSONAS` such as `System.out.println` and `Console.WriteLine`. Those patterns could not match the lowercased prompt text.

After fixing that, a second pipeline drift showed up: `compile_text()` always seeds IR with a default role such as `Helpful generative AI assistant`, while `DomainHandler.handle()` only applied implied personas when the role was empty. That meant handler tests passed, but `compile_text_v2()` and `/compile` still returned the generic role with no `metadata.implied_persona`.

## Validation
- `python -m pytest tests/test_regex_precompilation.py tests/test_domain_confidence.py tests/test_compile_policy_api.py -q`
- `python -m ruff check app/heuristics/handlers/domain_expert.py tests/test_domain_confidence.py tests/test_compile_policy_api.py`
- `python -m ruff format --check app/heuristics/handlers/domain_expert.py tests/test_domain_confidence.py tests/test_compile_policy_api.py`
- `git diff --cached --check`


madara88645:
Closing this stale/conflict PR because the same implied-persona fix has been rebuilt on top of the current main in #389. Please track the fresh implementation there.

PR #512, opened 2026-05-03
[codex] Harden Fly memory defaults and move frontend auth behind proxy

madara88645:
## What changed

- remove eager backend compiler initialization so Fly machines can answer lightweight health requests without loading the heavy compiler on startup
- add request-level backend logging with path, status, duration, client IP, user agent, and API key owner metadata
- raise Fly machine defaults from 256MB to 512MB and tighten request concurrency defaults
- add Next.js same-origin proxy routes for protected generator and RAG endpoints so the browser no longer needs to send a backend API key directly
- stop injecting `NEXT_PUBLIC_API_KEY` into browser requests and update frontend contract tests around same-origin routing
- document the new `PROMPTC_SERVER_API_KEY` flow and add operational regression coverage
- add a `.dockerignore` to shrink Fly deploy context size

## Why this changed

Fly emailed an OOM alert after `uvicorn` was killed on a 256MB machine. At the same time, the frontend was still capable of forwarding a public `NEXT_PUBLIC_API_KEY` from the browser to protected backend routes. This PR addresses both problems together:

- reduce startup memory pressure and improve observability
- move protected frontend auth behind a server-side proxy
- prepare backend CORS to be tightened safely after the frontend deploy lands

## Impact

- production Fly machines get safer default memory and concurrency settings
- protected frontend flows now expect `PROMPTC_SERVER_API_KEY` on the Next.js server instead of `NEXT_PUBLIC_API_KEY` in the browser
- same-origin proxy routes let us remove wildcard CORS from Fly without breaking the web app
- request logs become much more useful for tracing sudden traffic or resource spikes

## Validation

- `python -m pytest tests/test_operational_hardening.py tests/test_api_hardening.py tests/test_auth_fast_path.py -q`
- `cd web && npm run test:contracts`
- `cd web && npm run test`
- `cd web && npm run build`

## Follow-up

- Vercel production deploy still needs to land before removing the old `NEXT_PUBLIC_API_KEY` env and tightening Fly CORS fully in production
- a direct `vercel deploy` attempt from the CLI hit the account upload-rate limit (`api-upload-free`), so GitHub-triggered deploys are the intended path here


madara88645:
Holding this PR before merge because the overall direction is right, but the current proxy implementation is risky for our production constraints. The backend runs on Fly with tight RAM limits, so buffering the full request body and the full upstream response in memory is something I do not want to ship yet, especially around larger RAG-style payloads.\n\nBefore merge, I want to see three things:\n1. Stream request/response bodies through the proxy instead of materializing both sides fully in memory.\n2. Build the backend target URL safely (avoid raw string concatenation so trailing-slash configs cannot produce malformed paths).\n3. Align the server API key behavior and messaging: either remove the ADMIN_API_KEY fallback here, or document/error-message it clearly if the fallback is intentional.\n\nOnce those are handled, I am happy to re-check and likely merge this.

madara88645:
Bu PR #523 tarafından supersede edildi (main'e merge edildi). Ancak orijinal 3 gereksinim #523'te de tam karşılanmadı — şu an production'da aktif sorunlar:

1. **Streaming eksik**: `backendProxy.ts` hâlâ `arrayBuffer()` ile hem request hem response'u tamamen belleğe alıyor. Fly 512MB sınırında büyük RAG payload'larında risk.
2. **Güvenli URL build**: ``${resolveBackendApiBase()}${backendPath}`` — trailing slash kontrolü yok, malformed path riski var.
3. **ADMIN_API_KEY fallback**: Kaldırılmadı, belgelenmedi. Sessiz fallback davranışı.

Bu 3 madde için yeni Jules görevleri açılması önerilir.

PR #514, opened 2026-05-04
[codex] add same-origin backend proxy routes

madara88645:
## Summary
- add same-origin Next route proxies for compile, generator, benchmark, optimize, and RAG backend calls
- move protected browser flows to server-side API key forwarding via `PROMPTC_SERVER_API_KEY`
- add contract coverage for proxy route wiring and backend proxy behavior

## Why
The frontend was still depending on browser-visible backend routing and generator API key behavior. This PR moves protected calls behind Next server routes and adds fast regression coverage so proxy path mapping and auth behavior stay locked in CI.

## Validation
- `cd web && npm run test:contracts`
- `cd web && npm run build`
- `cd web && npm run lint`


madara88645:
Closing this in favor of #512. The proxy direction makes sense, but I do not want to split follow-up fixes across two overlapping PRs. The remaining merge blockers here are the same class of concerns I would rather solve once in the stronger branch: memory usage under low-RAM Fly constraints, safer backend URL construction, and final proxy contract cleanup.\n\nIf we continue this work, let us land it through #512 after the RAM-sensitive proxy behavior is tightened.

PR #546, opened 2026-05-09
[codex] Harden agent packs manifest contract

madara88645:
## Summary
- tighten the Agent Packs manifest model so `provider`, `kind`, and `preview_order` only accept the Claude values the UI expects
- reject manifests whose `preview_order` references groups that do not exist in the returned files
- add regression tests covering invalid provider, invalid file kinds, and stale preview groups

## Why
Recent Agent Packs work added a new backend/frontend contract, but the backend model still accepted arbitrary strings for manifest fields that the frontend treats as a fixed set. That made drift easy to introduce silently and hard to catch before runtime.

## Impact
This makes contract drift fail fast during tests and model validation instead of surfacing later in the UI.

## Validation
- `python -m pytest tests/test_agent_packs_api.py -q`

### DIFF
--- a/app/adapters/agent_packs.py
+++ b/app/adapters/agent_packs.py
@@ -5,7 +5,7 @@ from typing import Any, Literal, Protocol
 from zipfile import ZIP_DEFLATED, ZipFile
 
 from fastapi.responses import Response
-from pydantic import BaseModel, Field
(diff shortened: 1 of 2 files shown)


madara88645:
Superseded by #569. The manifest contract hardening was salvaged there on top of current main with passing verification.

PR #552, opened 2026-05-10
feat: add repo-aware public GitHub generator MVP

madara88645:
## Summary
- add a shallow public GitHub repo analysis endpoint for generator workflows
- pass optional `repo_context` through agent/skill generation and show a repo brief preview in the UI
- harden the Next proxy + generator pages so protected analyze requests do not get stuck in a perpetual loading state

## Testing
- `python -m pytest tests/test_repo_context.py tests/test_context_generation.py tests/test_skills_generator.py tests/test_api_hardening.py -q`
- `cd web && npm run test -- app/agent-generator/page.test.tsx app/skills-generator/page.test.tsx app/proxy-routes.test.ts lib/server/backendProxy.test.ts lib/promise/withTimeout.test.ts`
- `cd web && npm run build`

## Notes
- this is intentionally a lightweight MVP: no local clone, no recursive source scan, no branch/subdir/file URL support, and no separate LLM-powered repo analysis step

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -25,6 +25,16 @@ Prompt Compiler is a FastAPI + Next.js product that turns vague requests into st
 - Frontend tests: `cd web && npm run test`
 - Frontend build: `cd web && npm run build`
 
+## Server-side environment variables
+The Next-side proxy and the backend each need their own keys; keep them out of the bundled JS.
+
+| Env var | Side | Purpose |
+| --- | --- | --- |
+| `PROMPTC_SERVER_API_KEY` | Next.js | Forwarded as `x-api-key` to protected backend routes (generators, analyze). Without it, protected proxy routes return 500. |
+| `PROMPTC_PROXY_UPSTREAM_TIMEOUT_MS` | Next.js | Hard upstream-fetch timeout for the proxy (default 25000). Aborts a stuck backend connection with a 504 instead of hanging the route forever. |
+| `PROMPTC_GITHUB_TOKEN` (or `GITHUB_TOKEN`) | Backend | Optional. When set, the public-repo analyzer adds `Authorization: Bearer <token>` to GitHub requests, raising the rate limit from 60 req/h (anonymous) to 5000 req/h. |
+| `PROMPTC_REPO_CONTEXT_CACHE_TTL` | Backend | Repo-brief cache TTL in seconds (default 600). Set to `0` to disable the in-memory cache. |
+
 ## Domain Concepts
 - Conservative mode should avoid hallucinated requirements and fake APIs.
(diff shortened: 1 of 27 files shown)


madara88645:
Implementation notes / known gaps for reviewers:

1. Generation quality is improved by repo awareness, but it is still only MVP-grade. In manual testing, the generated agent/skill outputs can remain somewhat template-heavy and may still invent tool names or capability labels unless the upstream generator prompts are tightened further.

2. The repo analyzer is intentionally shallow. It only uses GitHub metadata, the root README, selected manifest files, and a very small amount of top-level directory structure. That keeps context small, but it also means the generated brief can miss important architectural conventions, internal abstractions, or monorepo boundaries.

3. The implementation currently depends on unauthenticated GitHub API access. That is fine for an MVP, but it leaves us exposed to GitHub rate limits and transient upstream failures, especially if multiple users analyze public repos from the same deployment/IP.

4. There is no caching layer yet. Re-analyzing the same repo repeats the full metadata/README/manifest fetch path, so latency and GitHub API usage are higher than they need to be.

5. URL support is deliberately narrow. Only `https://github.com/owner/repo` root URLs are accepted. Branch URLs, subdirectory URLs, blob URLs, and `.git` URLs are rejected. That matches the MVP scope, but reviewers should expect follow-up product pressure here.

6. The client-side stuck-loading mitigation is a timeout-based safety net, not a fundamental guarantee that every browser/runtime edge case has been eliminated. It prevents an infinite spinner, but a deeper end-to-end investigation may still be worthwhile if we see more reports from the in-app browser or other embedded runtimes.

7. The repo brief summary is bounded, but still dense. For smaller context-window models, we may need a second pass that compresses or prioritizes the brief even more aggressively before it is injected into generation prompts.

8. The analyzer intentionally avoids source-level reading. That reduces risk and token cost, but it also means the generated output should not be treated as code-audit quality guidance. Reviewers should evaluate whether the current prompt wording makes that limitation clear enough.

9. Local web testing still depends on correct server-side proxy configuration (`PROMPTC_SERVER_API_KEY` on the Next side and a valid backend key on the API side). This PR improves the failure path, but it does not remove that operational requirement.

10. I did not widen verification to the full repository test matrix from AGENTS.md. The PR was validated with targeted backend tests, targeted frontend tests, and a frontend production build. That gives good confidence for this slice, but not full-repo certainty.

Recommended follow-ups after merge:
- tighten generator prompts to explicitly forbid invented tools/APIs/sections
- add caching for analyzed repo briefs
- consider authenticated GitHub requests in hosted environments
- consider a smaller, model-budget-aware repo brief mode for weaker models
- consider telemetry on analyze failures/timeouts so we can see whether the browser pending issue fully disappears

PR #605, opened 2026-05-20
feat(web): implement seamless workflows, local optimizer fallback, and benchmark UX enhancements

madara88645:
This PR resolves 3 key UX and workflow challenges identified in the first-time user experience walkthrough across the Web Compiler, Token Optimizer, and Benchmark pages.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,7 +8,7 @@ LLM_SKILL_MAX_TOKENS=2048
 # Compiler behavior
 PROMPT_COMPILER_MODE=conservative
 
-# Optional auth hardening
+# Optional internal auth hardening (public app routes do not require end-user Prompt Compiler API keys)
 ADMIN_API_KEY=
 PROMPTC_REQUIRE_API_KEY_FOR_ALL=false
 
@@ -20,4 +20,3 @@ PROMPTC_RAG_ALLOWED_ROOTS=
 
 # Optional frontend override
 NEXT_PUBLIC_API_URL=http://127.0.0.1:8080
-PROMPTC_SERVER_API_KEY=
--- a/.jules/instructions.md
+++ b/.jules/instructions.md
@@ -19,7 +19,8 @@ Treat journal entries as **context**, not tasks. Do not refactor unrelated modul
 2. **Secrets:** Never expose `.env`, credentials, or database URLs in code, logs, or PR text.
 3. **Generated artifacts:** Treat emitted agent packs, skills, and MCP stubs as **untrusted until reviewed** (see CLAUDE.md Security).
 4. **Conservative mode:** Do not invent APIs or requirements when conservative / policy-aware paths apply.
-5. **Scope:** Focused tests before full suites (`python -m pytest tests/ -q` or targeted files per CLAUDE.md).
+5. **Public product auth:** Do not introduce end-user Prompt Compiler API key requirements, browser API-key inputs, or hidden proxy-secret assumptions for public web/app routes. Open-source users should be able to use the product without learning `x-api-key`, `PROMPTC_SERVER_API_KEY`, or similar internal knobs.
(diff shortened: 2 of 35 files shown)


madara88645:
This PR is larger than it may look at first glance, so here is the intended merge context and why it matters.

## What this PR is doing

The main purpose of this PR is to remove **end-user Prompt Compiler API key requirements** from the public/open-source product experience.

In practice, that means:

- Public-facing app flows should **not** ask visitors for internal Prompt Compiler keys such as `x-api-key`, `PROMPTC_SERVER_API_KEY`, or similar custom auth values.
- The app should behave like an open-source product that people can open and use, not like a private developer API that expects undocumented internal credentials.
- Provider credentials such as `OPENAI_API_KEY` / `GROQ_API_KEY` remain a **server/operator concern**, not something the end user is expected to know or supply in the UI.

## Key changes included here

- Removed custom Prompt Compiler API-key gating from the intended public routes and flows.
- Updated frontend messaging so users are no longer told to provide undocumented internal keys.
- Updated repository instructions/docs/agent guidance so future agents do not silently reintroduce this requirement.
- Fixed related CI/smoke/lint issues uncovered while making the auth surface consistent.
- Added an offline fallback for `/validate` so public validation does not crash in environments where no upstream LLM key is present.
- Resolved merge conflicts with `main` while preserving newer safety/error-handling improvements from upstream.

## Why this PR is important

This is not just a small auth tweak. It corrects a product-level mismatch.

Before this change, the repo had multiple places where the system behaved as if a random visitor should already know:

- what `x-api-key` is,
- what `PROMPTC_SERVER_API_KEY` is,
- where such a key would come from,
- and why they would need it in the first place.

For an open-source/publicly usable product, that is a major UX and adoption problem.

The previous behavior created confusion because the application exposed **internal protection concepts** to end users without giving them any realistic path to succeed. In other words, people could reach the product, try it, and then get blocked by a requirement that was neither documented for them nor appropriate for them.

This PR intentionally removes that mismatch.

## Why this should be reviewed carefully before merge

This PR changes an important boundary: **who authentication is for**.

The explicit decision behind this branch is:

- Do **not** use custom Prompt Compiler API keys as a requirement for normal public product usage.
- If stronger protection is needed later, it should be added through the appropriate mechanisms for the deployment model (for example: rate limits, quotas, login, admin-only controls, infrastructure-level protections, or other operator-owned safeguards).
- Do **not** push that burden onto anonymous users through undocumented key prompts.

So if anyone reviewing this PR expected some of these routes to remain protected by end-user-entered Prompt Compiler keys, that would be a product-direction disagreement, not just an implementation detail.

## Merge intent

Please merge this PR with the following intent in mind:

1. Public/open-source users should be able to use the product without learning internal Prompt Compiler auth knobs.
2. Internal/server/provider credentials should stay internal.
3. Future agents and contributors should treat reintroducing browser-side Prompt Compiler key prompts as a regression unless there is an explicit product decision to change direction.

## Verification notes

The branch was updated to keep smoke/lint/test expectations aligned with this behavior, including the `/validate` offline fallback so CI does not fail just because provider credentials are absent.

If you are merging this with an AI agent later, the most important thing for the agent to preserve is the policy intent above, not just the literal code diff.

madara88645:
Deferring this PR for human review rather than auto-merging in the current batch session. CI is green and the PR is `MERGEABLE`, but the diff bundles three concerns (seamless workflows, local optimizer fallback, benchmark UX) across 34 files and includes a 246-line deletion of `web/lib/server/backendProxy.ts` + its test file. The branch name (`chore/jules-instructions-guardrails`) also doesn't match the current title (`feat(web): …`), suggesting the branch was repurposed. Could you split this into ≤3 focused PRs (one per concern), or post a brief justification for the `backendProxy` removal so it can be reviewed on its own merits?

PR #626, opened 2026-05-24
⚡ Bolt: Optimize vector math operations in RAG indexing

madara88645:
💡 What: Replaced the `v2.get(k, missing)` pattern with a native `if k in v2:` check for sparse dictionary dot products in `cosine_similarity`.
🎯 Why: Iterating over the smaller dictionary and checking keys with the native `in` operator is faster than using `.get()` with a sentinel object, improving the performance of semantic chunking during text ingestion.
📊 Impact: Expected to reduce CPU overhead during heavy text ingestion when computing TF-IDF cosine similarities.
🔬 Measurement: Run the backend test suite `python -m pytest tests/test_rag.py` to ensure RAG index functionality remains perfectly intact.

---
*PR created automatically by Jules for task [17302220293516765569](https://jules.google.com/task/17302220293516765569) started by @madara88645*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -8,7 +8,6 @@
 **Learning:** In the Simple Index RAG module (`app/rag/simple_index.py`), calculating the TF-IDF for sentences happens iteratively within an inner loop inside `compute_tfidf` function which is called for each sentence against a `doc_freq` Counter. Re-calculating the IDF via `math.log` for the entire dataset inside the inner loop was an unnecessary computation overhead. Since the document count and total frequency mapping are static, these metrics can be pre-calculated upfront using a dictionary cache to allow O(1) lookups.
 **Action:** Precompute static mapping dependencies inside tight inner loops with caching logic to mitigate unnecessary redundant execution overhead.
 
-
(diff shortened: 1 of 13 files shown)


madara88645:
This branch now conflicts with main after #642 landed the SQL placeholder change in app/rag/simple_index.py. Two requests before we can move forward: (1) please rebase against main; (2) this PR is 13 files / +117-26 with no review attached — could you either split it into smaller logical PRs, or add a short benchmark (before/after) confirming the cosine_similarity optimization actually reduces CPU during ingestion? Either makes it much easier to review safely.

madara88645:
Closing as superseded — the proposed `cosine_similarity` change (sparse-vector `if k in v2` direct lookup instead of `.get(k, missing)` sentinel) already exists on `main` at `app/rag/simple_index.py:389-393` with the same explanatory comment about lower bytecode overhead. The branch is also ~267 lines behind main on 25 unrelated files (CI, frontend, backend hardening), so a rebase-and-merge would revert that work. Thanks Bolt — pattern accepted, just via a different path.

PR #676, opened 2026-05-29
🎨 Palette: [UX improvement] Add example buttons to generative empty states

madara88645:
**💡 What:** 
Added "or try an example" helper buttons below the primary action buttons on all generative tool pages (`offline`, `optimizer`, `benchmark`, `agent-generator`, `skills-generator`, and `agent-packs`). When the respective input text areas are empty, these buttons are visible and allow users to instantly populate the input with a contextually appropriate, high-quality sample prompt.

**🎯 Why:** 
The "blank canvas" problem is a significant UX hurdle, particularly in generative AI tools where users may not immediately know the expected format, length, or level of detail required for a prompt. While the main `/page.tsx` compiler page successfully utilized this pattern, other specialized tools lacked it, leaving users to guess. This enhancement makes the entire suite of tools immediately approachable and teaches users the expected input formats by example.

**📸 Before/After:** 
_Before:_ The empty states only showed a disabled primary action button.
_After:_ A subtle "or try an example" button appears, which populates the text area and focuses it, enabling immediate interaction. (Visual verification attached in verification logs).

**♿ Accessibility:** 
- The buttons use the native HTML `<button>` element with `type="button"` and are fully keyboard-navigable via the `Tab` key.
- Focus states are explicitly styled using `focus-visible:ring-1` to support visual keyboard navigation.
- Clicking the button programmatically sets focus to the respective `textarea`, ensuring screen reader and keyboard users maintain context and can immediately begin editing the sample prompt if desired.

_Also includes a minor code health fix for an ESLint `react-hooks/set-state-in-effect` error in `useContextManager.ts`._

---
*PR created automatically by Jules for task [11622838760355911749](https://jules.google.com/task/11622838760355911749) started by @madara88645*


madara88645:
Not merged in this batch: branch has conflicts with `main`. Please rebase onto the latest `main`.

madara88645:
Closing as stale. This branch was cut from an older `main` and now diffs against ~20-40 files of unrelated work that has since merged, instead of the small change described in the title. Merging it would risk reverting good work on `main`, and a rebase would be dominated by that divergence rather than the intended fix.

The change itself still looks worthwhile — please regenerate it as a **fresh, scoped PR against current `main`** (ideally touching only the file(s) named in the title). That will be far easier to review and merge.

PR #753, opened 2026-06-05
🛡️ Sentinel: [HIGH] Fix Command Injection in Editor Invocation

madara88645:
🚨 **Severity:** HIGH
💡 **Vulnerability:** Command Injection (CWE-77). The application launched the user's preferred text editor defined via the `$EDITOR` or `$VISUAL` environment variables using `subprocess.run()`. It attempted to protect against arbitrary command execution using a brittle denylist of executable names (`bash`, `python`, etc) and shell metacharacters. This manual parsing and filtering is inherently flawed and bypassable via non-blocked execution wrappers or path-manipulation tricks.
🎯 **Impact:** If an attacker could influence the environment variables, or if a user accidentally had a maliciously formed string in their `$EDITOR` configuration, arbitrary code could be executed when attempting to quick-edit a prompt.
🔧 **Fix:** Replaced the manual parsing, temp file handling, and `subprocess.run()` logic with `click.edit(text, require_save=True)`.
- **Windows Path Handling:** Kept the Windows-specific path formatting to quote paths containing spaces correctly.
- **Specific Exception Handling:** Catch `click.ClickException` specifically for cleaner warnings and generic `Exception` for unexpected crashes, returning `None` in both failure scenarios.
- **Cancel/Failure Handling:** Return `None` when user cancels or when editing fails, rather than the original prompt text.
✅ **Verification:**
- Added non-mocked malicious `EDITOR`/`VISUAL` validation tests proving blocked values never reach the execution boundary (e.g., `bash -c`, `cmd /c`, `powershell -Command`).
- Added non-mocked safe temporary editor integration test using a temporary python script acting as an editor.
- Kept the existing Windows quoted-path test.
- Ran all CLI and quick-edit tests successfully, and all CI check suites are passing.

### DIFF
--- a/app/quick_edit.py
+++ b/app/quick_edit.py
@@ -4,6 +4,8 @@ This module provides functionality to quickly edit prompts from history and favo
 including text editing, metadata updates, and re-compilation.
 """
 
+import os
+import shlex
 import click
 from typing import Any, Dict, Literal, Optional
 
@@ -54,6 +56,7 @@ FORBIDDEN_EDITOR_PREFIXES = (
     "perl",
     "php",
 )
+SHELL_METACHAR_TOKENS = {";", "|", "&", "&&", "||"}
 
 
 class QuickEditor:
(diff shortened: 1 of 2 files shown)


madara88645:
Thanks for tackling a risky area. After reviewing the actual boundary, this should not be merged as a security fix in its current form. Switching to click.edit does simplify the code, but it still reaches process execution through the configured EDITOR/VISUAL path, so the underlying trust boundary is not actually proven safer yet. Recommended next step: either narrow the claim and present this as a refactor, or add real validation plus direct regression coverage for malicious editor strings, wrapper binaries, quoted Windows paths, and preserved legitimate editor behavior before calling this a security hardening change.

madara88645:
REQUEST_CHANGES\n\nDo not merge this PR in its current form.\n\nThe claimed security boundary is not proven. Replacing the custom subprocess logic with \click.edit\ is not sufficient evidence that malicious \EDITOR\ / \VISUAL\ values are safely handled. This change simplifies the implementation, but it does not currently prove that the risky execution path is hardened.\n\nRequired before any future merge:\n- Add direct malicious \EDITOR\ / \VISUAL\ regression tests.\n- Prove the boundary against malicious editor strings, wrapper binaries, and quoted Windows paths.\n- Preserve legitimate editor behavior while demonstrating that unsafe values are rejected or otherwise safely contained.\n\nUntil that evidence exists, this should not be merged as a security fix.\n\nNote: a formal GitHub \REQUEST_CHANGES\ review could not be submitted here because GitHub does not allow requesting changes on your own pull request via the API.

madara88645:
Hold for dedicated security review. This does not provide enough new security value during launch. A clean replacement should prove the real EDITOR/VISUAL boundary with non-mocked tests.

madara88645 (commented):
REQUEST CHANGES: This PR is the better candidate in the editor-invocation security cluster:
- safer failure behavior: returns None on cancel/failure
- better Windows path handling
- keeps the denylist/validation concept

But it cannot be merged yet because the security boundary is still not proven. The current tests mock too much of the real execution path.

Required before merge:
1. Add non-mocked malicious EDITOR/VISUAL tests proving blocked values never reach process execution:
   - bash -c ...
   - sh -c ...
   - python -c ...
   - cmd /c ...
   - powershell -Command ...
2. Add a non-mocked safe editor integration test:
   - create a temporary Python script that behaves like an editor
   - pass it as EDITOR/VISUAL
   - prove the temp file is edited successfully
3. Confirm forbidden editor prefixes / denylist validation is not removed.
4. Confirm click.edit does not weaken the boundary compared with current subprocess list execution.
5. Preserve failure behavior: cancel/failure returns None, not original text.

Do not merge until these tests are added and CI is green.

PR #754, opened 2026-06-05
🎨 Palette: Add aria-pressed to ExportPanel toggle buttons

madara88645:
💡 What: Added `aria-pressed` attributes to the target and output mode toggle buttons in both `agent-generator` and `skills-generator` `ExportPanel` components.
🎯 Why: To explicitly convey the active/selected state of these mutually exclusive toggle buttons to screen readers, improving accessibility for visually impaired users.
📸 Before/After: Visual presentation remains identical, but DOM now includes `aria-pressed="true"` for active tabs.
♿ Accessibility: Ensures WCAG compliance for toggle button states.

---
*PR created automatically by Jules for task [1839656788384490074](https://jules.google.com/task/1839656788384490074) started by @madara88645*

### DIFF
--- a/web/app/agent-generator/components/ExportPanel.tsx
+++ b/web/app/agent-generator/components/ExportPanel.tsx
@@ -238,6 +238,7 @@ export default function ExportPanel({ systemPrompt, isMultiAgent }: ExportPanelP
                 type="button"
                 key={item.id}
                 onClick={() => handleTargetClick(item.id)}
+                aria-pressed={target === item.id}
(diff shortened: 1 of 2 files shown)


madara88645:
Thanks for the accessibility improvement here. The code change itself looks safe, but this PR should stay on hold until the failing VS Code Extension check is green again. The current failure looks like an integration-test/runtime issue rather than a product regression, but we should still rerun or stabilize that check before merging so launch hardening stays clean.

PR #794, opened 2026-06-11
⚡ Bolt: [performance improvement] Persistent HTTP client for GitHub API

madara88645:
💡 What: Replaced ephemeral `httpx.Client` context managers with a global, persistent HTTP connection pool for GitHub API requests. The client is safely closed during the FastAPI app shutdown lifespan event.

🎯 Why: In `_fetch_public_github_repo_payload`, a new `httpx.Client` was being initialized and torn down for every GitHub API request block. This incurs significant overhead because it forces the application to establish a new TCP connection and perform a fresh TLS handshake on every repository lookup.

📊 Impact: Reduces GitHub metadata fetch time by ~30% (from ~1.03s down to ~0.73s across 3 consecutive calls locally), greatly minimizing overhead and connection churn.

🔬 Measurement: Local tests (`test_perf.py`) show a measurable decrease in execution time across sequential requests. The entire backend test suite and linter pass successfully.

---
*PR created automatically by Jules for task [12816249187286437144](https://jules.google.com/task/12816249187286437144) started by @madara88645*

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -8,6 +8,7 @@ from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 
 from app import get_build_info
+from app.github_repo_context import close_github_http_client
 
 from api.shared import is_meta_leaked, logger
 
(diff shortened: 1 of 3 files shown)


madara88645 (commented):
REQUEST CHANGES: The global GitHub HTTP client must not persist Authorization headers across requests. Do not use client.headers.update(...) for per-request auth. Pass headers per request instead. Add a regression test proving a token from one request is not sent on the following tokenless request.

PR #828, opened 2026-06-23
test: pure utility helpers in app/emitters.py have no test coverage

madara88645:
## Summary

Adds `tests/test_emitters_utils.py` with unit tests for four pure helper functions in `app/emitters.py` that had zero coverage. These functions drive the conservative-mode greeting path and domain-suggestion deduplication logic.

### Functions under test

All four are side-effect-free (no env reads, no I/O, no model instantiation):

| Function | Role |
|---|---|
| `_is_trivial_input(text, domain, complexity)` | Gates conservative-mode template bypass for short/greeting inputs |
| `_contains_any_marker(text, markers)` | Fast-path substring check used for Turkish language detection |
| `_minimal_greeting_prompt(text, lang)` | Returns language-appropriate inline prompt for trivial inputs |
| `_clean_domain_suggestion_text(text)` | Strips short prefixes (≤8 chars) before named action markers |

### Test classes

**`TestIsTrivialInput`** (8 tests) — boundary conditions on the 30-char length gate, domain check, complexity values (`"low"`, `None`, `""`), and whitespace stripping before length measurement.

**`TestContainsAnyMarker`** (7 tests) — matching/non-matching markers, empty text, empty marker tuple, case sensitivity (no match on `"MERHABA"` vs `"merhaba"`), partial substring match.

**`TestMinimalGreetingPrompt`** (7 tests) — English vs Turkish output routing, empty-text defaults (`"hello"` / `"merhaba"`), whitespace collapsing, Turkish-marker detection when `lang="en"`, `"slm"` shorthand trigger.

**`TestCleanDomainSuggestionText`** (8 tests) — no-prefix pass-through, short-prefix stripping (`"Tip:"`, `"A:"`, `"1."`, `"Ok:"`), long-prefix not stripped (position > 8), empty string, whitespace normalization.

### Test count
30 tests total.

## Test plan
- [ ] `python -m pytest tests/test_emitters_utils.py -v` passes with 0 failures
- [ ] No source files modified — only new test file added
- [ ] No `.env`, config, lockfile, or CI changes

---
_Generated by [Claude Code](https://claude.ai/code/session_01LPYz7KruCnV4om157tWdfb)_

### DIFF
--- a/tests/test_emitters_utils.py
+++ b/tests/test_emitters_utils.py
@@ -6,8 +6,6 @@ no model imports) and are safe to call in isolation.
 
 from __future__ import annotations
 
-import pytest
-
 from app.emitters import (
     _clean_domain_suggestion_text,
     _contains_any_marker,
@@ -131,7 +129,7 @@ class TestCleanDomainSuggestionText:
 
     def test_whitespace_normalized(self) -> None:
         result = _clean_domain_suggestion_text("  Review   the code  ")
-        assert result == "Review   the code"
+        assert result == "Review the code"
 
     def test_ok_prefix_stripped_before_use(self) -> None:
         assert _clean_domain_suggestion_text("Ok: Use type hints") == "Use type hints"
--- a/tests/test_emitters_utils.py
+++ b/tests/test_emitters_utils.py
@@ -1,0 +1,137 @@
+"""Unit tests for pure utility helpers in app/emitters.py.
+
+Functions under test have no external dependencies (no env reads, no I/O,
+no model imports) and are safe to call in isolation.
+"""
+
+from __future__ import annotations
+
+import pytest
+
+from app.emitters import (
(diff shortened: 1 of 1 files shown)


madara88645:
## Cleanup fix pushed ✅

Found two issues while triaging:
1. **Unused `import pytest`** — the pre-commit/Smoke failure (ruff flagged it).
2. **A genuinely failing test** that CI hadn't surfaced (Smoke failed first, so the full suite was skipped): `test_whitespace_normalized` asserted `"Review   the code"`, but `_clean_domain_suggestion_text` **collapses all internal whitespace** (`"  a   b  "` → `"a b"`), so the correct expected value is `"Review the code"`.

Pushed a fix for both. Now **30 passed**, ruff + format clean. Marking ready to merge once CI is green.

PR #932, opened 2026-07-03
feat(web): typewriter animation parity for "try an example" across generators

madara88645:
## What
The typewriter "or try an example" effect previously existed **only** on the main compile page. This extracts it into a shared hook and applies it to every single-field generator surface.

- **New** \`web/app/hooks/useTypewriterFill.ts\` — one implementation of the effect (interval-based typing, \`prefers-reduced-motion\` bailout, focus, unmount cleanup, cancellable).
- **Refactor** \`web/app/page.tsx\` to consume the hook (removes the duplicated inline logic).
- **Applied to 6 surfaces:** agent-generator, skills-generator, benchmark, optimizer, agent-packs, offline (both example buttons). Each now types the example in and cancels an in-flight run on manual edit / clear.

Out of scope: \`pr-safety\` (multi-field example loader — doesn't fit the single-field pattern).

## Why
Consistent, polished autofill across all generators; single source of truth instead of one-off copies.

## Testing
- Hook unit tests: reduced-motion instant fill, animated typing over the interval, focus (3/3).
- Main-page regression test for example fill.
- Full web suite: **30 files / 161 tests pass**.
- \`npm run build\` (typecheck) passes.

## Notes
- No new dependencies (plain \`setInterval\` + React state).
- Honors \`prefers-reduced-motion\` on every surface (instant fill).
- Design + plan docs included under \`docs/superpowers/\`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #963, opened 2026-07-07
feat(web): auto-compile after typewriter example fill

madara88645:
## Summary
- The empty-state "or try an example" button previously only typewrote a sample prompt into the textarea, leaving the visitor to find and click Compile themselves.
- `fillExample` now takes an `autoCompile` flag; the example button passes `true`, so after the typewriter finishes (or immediately under `prefers-reduced-motion`) it calls the same `runCompile(text, activeMode)` path the Compile button uses.
- Manual typing into the textarea and the Compile button are unchanged — `autoCompile` defaults to `false` and is only opted into by the example entry point.

## Notes on overlapping in-flight work
- PR #932 ("typewriter animation parity") is open and also touches this `fillExample` flow (extracts it into a `useTypewriterFill` hook). Since it isn't merged, this PR is built against the current `main` implementation of `fillExample` and will likely need a rebase after #932 merges.
- A parallel task (T10) is turning the empty-state italic use-case text into clickable chips that also call `fillExample()`, in the same file/adjacent region — expect a possible merge conflict with that PR too.

## Test plan
- [x] `cd web && npm run test` — 33 test files / 172 tests pass, including a new `web/app/__tests__/page-example-autocompile.test.tsx` covering: auto-compile on reduced motion, auto-compile after the typewriter interval finishes, and no auto-compile on manual typing (Compile button still required)

### DIFF
--- a/web/app/__tests__/page-example-autocompile.test.tsx
+++ b/web/app/__tests__/page-example-autocompile.test.tsx
@@ -1,0 +1,111 @@
+import { act, fireEvent, render, screen } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import Home from "../page";
+
+const runCompileMock = vi.fn();
+const useCompilerMock = vi.fn();
+const useContextManagerMock = vi.fn();
+
+vi.mock("../hooks/useCompiler", () => ({
+  useCompiler: () => useCompilerMock(),
+}));
+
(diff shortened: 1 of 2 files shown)


madara88645:
Superseded by #972, which integrated the 8 mutually-conflicting page.tsx B-cluster PRs (this one included) into one conflict-resolved, tested change. Merged to main as ec00971f. Closing as its content is now on main.

PR #971, opened 2026-07-07
fix(web): stop offline page from claiming local-only/no API keys

madara88645:
## Summary
- The `/offline` page's InfoButton claimed it was "a fast, local-only version of the Prompt Compiler ... it's secure, instant, and requires no API keys." In reality it called the exact same backend over the network as the main Compiler page — the only functional difference was `v2: false` in `useCompiler`/`compilePrompt`, which skips the LLM call and stays on the server's deterministic heuristic pipeline. That claim was misleading and the page was also duplicating the main Compiler's job.
- **Approach taken: fold-in (preferred option), not the copy-rewrite fallback.** I tried the preferred approach first and it worked cleanly:
  - Added a "Heuristics only (no LLM)" engine toggle to the main Compiler page (`web/app/page.tsx`), next to the existing Conservative switch. It threads a new `useLlm` boolean through `useCompiler.ts` (`runCompile`/`retry`/`resolveSecurityDecision`), which maps directly to the existing `v2` request flag — no backend changes.
  - When the heuristic engine is active, the "AI MODE" badge now correctly reads "HEURISTIC MODE" instead of falsely claiming AI/LLM involvement, and the Conservative switch is disabled with an explanatory title, since `mode` only affects the LLM code path server-side (confirmed in `api/routes/compile.py::_compile_response`) and has no effect once heuristics-only is selected.
  - Removed the "Offline" entry from `Sidebar.tsx` (and its accent color map entries) since the surface no longer needs a standalone nav item.
  - `web/app/offline/page.tsx` is now a redirect: it sends visitors (old bookmarks/links) to `/` instead of duplicating the compiler UI, with a short explanatory message and a manual link as a JS-disabled fallback.
- Why fold-in over rename: the "no LLM" engine choice is genuinely just a request flag on the same page's flow (mode/conservative already lives there), so keeping a whole separate route+page in sync with the main page's tabs/output UI going forward was the real duplication risk, not a naming problem. Folding it in removes that maintenance burden entirely rather than just fixing the copy.
- Note: PR #932 also touches `web/app/offline/page.tsx` (typewriter refactor). Since this PR replaces that file's contents with a small redirect component, expect a conflict there — whichever of these two merges second will need a quick rebase/resolution (trivial: keep the redirect, drop the typewriter usage on this now-removed page).

## Test plan
- [x] `cd web && npm run test` — 33 test files / 174 tests pass (added `page-heuristics-toggle.test.tsx` and `offline-page-redirect.test.tsx`, replaced `offline-page-error-state.test.tsx`, updated `Sidebar.test.tsx`)
- [x] `cd web && npm run build` — production build succeeds; `/offline` still resolves as a page (now a redirect)
- No backend files were touched (`app/`, `api/` untouched, per task scope)


madara88645:
Superseded by #972, which integrated the 8 mutually-conflicting page.tsx B-cluster PRs (this one included) into one conflict-resolved, tested change. Merged to main as ec00971f. Closing as its content is now on main.

PR #974, opened 2026-07-07
feat(web): show destination path for single-file exports

madara88645:
## Summary
- Exporting a Claude Subagent (or a single-file Claude MCP Tool export) never told the user where the resulting file goes; the backend already returns a `path` (e.g. `.claude/agents/{slug}.md`, see `app/adapters/claude_code.py:103`) but the frontend only surfaced it once a result had more than one file.
- Both `ExportPanel.tsx` components (agent-generator and skills-generator) now render the file's path as a small monospace label directly above the code block whenever the export result has exactly one file, using the same `path` value already present on the file object. Multi-file behavior (the skills-generator file-select dropdown) is unchanged.
- Updated both `ExportPanel.test.tsx` files with a new test asserting the path is visible when there's exactly one file.

Note: PR #970 previously rewrote `web/app/agent-generator/components/ExportPanel.tsx` significantly — it is already merged into `main`, so this branch already includes those changes and no reconciliation is needed there.

Note: a parallel task (T23, download/zip buttons) touches the same two `ExportPanel` files in a nearby region (~line 274 area). Expect a small merge conflict when both land — both changes are additive (a path label vs. download buttons), so it should be trivial to reconcile.

## Test plan
- [x] `cd web && npm run test` — 38 test files / 194 tests passed (pre-existing happy-dom `AbortError` teardown console noise, unrelated to this change and not a failure)
- [x] `cd web && npm run build` — succeeds (Next.js/Turbopack build + TypeScript check)

### DIFF
--- a/web/app/agent-generator/components/ExportPanel.test.tsx
+++ b/web/app/agent-generator/components/ExportPanel.test.tsx
@@ -95,6 +95,27 @@ describe("Agent ExportPanel", () => {
     expect(JSON.parse(options.body).format).toBe("claude-agent-sdk-ts");
   });
 
+  test("shows the destination file path for a single-file export result", async () => {
+    apiFetch.mockResolvedValue({
+      ok: true,
+      json: async () => ({
+        python_code: null,
+        yaml_config: null,
+        code: "# Agent\n\nRole: Test",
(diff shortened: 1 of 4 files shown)


madara88645:
Superseded by #978, which integrated the 5 conflict-affected C-cluster PRs (this one included) into one conflict-resolved, tested change. Closing as its content is now on main.

PR #977, opened 2026-07-07
feat(web): per-file and ZIP download in export panels

madara88645:
## Summary
- Moved `buildPackZip`/`zipPackBytes` from `web/app/agent-packs/lib/packZip.ts` to the shared `web/app/lib/packZip.ts` so both export panels can reuse the same zipping logic as Agent Packs. The zip helper's parameter type was narrowed to a minimal `{ path, content }` shape instead of importing `AgentPackFile`, so it also directly accepts the lighter `ExportFile` shape used by the agent-generator/skills-generator export results without any extra mapping.
- Added a "Download file" button to both `web/app/agent-generator/components/ExportPanel.tsx` and `web/app/skills-generator/components/ExportPanel.tsx` that downloads the currently selected export as a blob, named by its file path's basename (falling back to a format+extension name when the export has no `files` array, e.g. raw Python/TypeScript/JSON code).
- Added a "Download .zip" button, shown only when the export result contains more than one file, that bundles all of them via the shared `buildPackZip`.
- Existing copy-to-clipboard behavior is unchanged — these are additive buttons alongside Copy in the same toolbar.

## Overlap notes
- PR #970 already rewrote `web/app/agent-generator/components/ExportPanel.tsx` to hand off the "Claude Project Pack" target to `/agent-packs` instead of an inline multi-file dropdown — that PR is merged on `main` already (commit `b41025f9`), so this PR was built directly against that current shape (no reconciliation needed there). Flagging in case any other still-open branch based on the pre-#970 shape needs a rebase.
- T24 (destination path label) touches the same two `ExportPanel.tsx` files — expect a merge conflict there, should be trivial to reconcile (adjacent lines, no logic overlap).

## Test plan
- [x] `cd web && npm run test` — 38 files / 196 tests passed (pre-existing `ECONNREFUSED` noise from unrelated tests hitting a local port, not caused by this change)
- [x] `cd web && npm run build` — Next.js build succeeds
- [x] New tests: single-file download triggers a blob download with the correct filename; ZIP download button only renders (and works) when the export result has more than one file, in both `agent-generator` and `skills-generator` `ExportPanel.test.tsx`

### DIFF
--- a/web/app/agent-generator/components/ExportPanel.test.tsx
+++ b/web/app/agent-generator/components/ExportPanel.test.tsx
@@ -1,5 +1,5 @@
 import { fireEvent, render, screen, waitFor } from "@testing-library/react";
-import { beforeEach, describe, expect, test, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
 
 import ExportPanel from "./ExportPanel";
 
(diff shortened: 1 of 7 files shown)


madara88645:
Superseded by #978, which integrated the 5 conflict-affected C-cluster PRs (this one included) into one conflict-resolved, tested change. Closing as its content is now on main.

PR #983, opened 2026-07-07
Fix duplicate React keys in RAG search results

madara88645:
## Summary
- RAG search returns chunk-level rows, so a single uploaded document split into multiple chunks can produce several results sharing the same `path` (see `api/routes/rag.py:130-140`).
- `RagSearchPanel` keyed each result card by `result.path` alone, so React collided/dropped cards for same-path chunks — hitting the most common first-session RAG flow (search one uploaded doc).
- Fix: key each card by `${result.path}-${index}` so every chunk row renders distinctly, even when paths repeat.

## Test plan
- [x] Extended `web/app/components/__tests__/RagSearchPanel.test.tsx` with a case rendering two results sharing the same path, asserting both snippets render and the path label appears twice.
- [x] `cd web && npm run test` — 213/213 passing (41 test files).
- [x] `cd web && npm run build` — succeeds.


madara88645:

PR #984, opened 2026-07-07
Humanize export errors and add retry in ExportPanel

madara88645:
## Summary
- Both the agent-generator and skills-generator ExportPanels rendered the raw fetch/catch error message (e.g. `Failed to fetch`) with no way to recover, unlike the rest of the app which already uses `describeRequestError`.
- Format the caught export error with `describeRequestError` (`web/config.ts`) so users see a friendly, actionable message instead of raw browser errors.
- Add a "Retry export" button inside the error box in both panels that re-calls the existing `fetchExport(target, outputMode)`, reusing the current cache/state — no change to caching semantics.

## Test plan
- [x] `cd web && npm run test` — 41 test files / 214 tests pass, including new friendly-error-message + retry-button coverage in both `ExportPanel.test.tsx` files

Scope note: only touched the error-display/fetch-catch regions in both `ExportPanel.tsx` files; no clipboard-copy call sites, `app/agent_packs_*.py`, or `cli/commands/agent_packs.py` were modified.

### DIFF
--- a/web/app/agent-generator/components/ExportPanel.test.tsx
+++ b/web/app/agent-generator/components/ExportPanel.test.tsx
@@ -11,9 +11,13 @@ const { push } = vi.hoisted(() => ({
   push: vi.fn(),
 }));
 
-vi.mock("@/config", () => ({
-  apiFetch,
-}));
+vi.mock("@/config", async () => {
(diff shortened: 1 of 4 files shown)


madara88645:

PR #985, opened 2026-07-07
fix(web): trap focus and support Escape in SecurityAlert modal

madara88645:
## Summary
- The security findings modal (`SecurityAlert.tsx`) renders `role=dialog aria-modal` with `autoFocus`, but had no Escape-key handler and no focus trap, so Tab could walk focus out into the obscured page behind it.
- Add a keydown effect that calls `onCancel` on Escape.
- Add a minimal first/last-focusable-element Tab loop (no new dependencies) so Tab/Shift+Tab stay contained within the modal.
- Add backdrop-click-to-cancel (clicking the dimmed overlay outside the dialog calls `onCancel`; clicks inside the dialog do not).
- `autoFocus` on the "Strip Secrets & Proceed" button is unchanged.

## Test plan
- [x] `cd web && npm run test` (216 tests pass, including 4 new `SecurityAlert.test.tsx` cases: Escape closes, backdrop click closes, click-inside does not close, Tab/Shift+Tab stay contained)
- [x] `cd web && npm run build`

### DIFF
--- a/web/app/components/SecurityAlert.test.tsx
+++ b/web/app/components/SecurityAlert.test.tsx
@@ -1,0 +1,89 @@
+import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, expect, test, vi } from "vitest";
+
+import SecurityAlert from "./SecurityAlert";
+import type { SecurityFinding } from "../../lib/api/types";
+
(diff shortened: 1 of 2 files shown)


madara88645:

PR #986, opened 2026-07-07
fix(web): guard Benchmark against double submission

madara88645:
## Summary
- Holding Cmd+Enter (or double-clicking Run Benchmark) on the Benchmark page had no guard, so it could fire parallel runs of the most expensive endpoint in the app (two LLM calls per run).
- Add an `isRunningRef` guard in `handleBenchmark` (set at start, cleared in `finally`, early-return if already set), mirroring the existing pattern in `web/app/pr-safety/page.tsx`'s `isAnalyzingRef`.
- Add `if (event.repeat) return;` to the prompt textarea's `onKeyDown`, matching the pattern already used in `web/app/agent-generator/page.tsx`.
- No UI changes — purely defensive.

## Scope
- Only touched `handleBenchmark` (submit handler) and the textarea `onKeyDown` in `web/app/benchmark/page.tsx`. Did not touch the results-row layout region (~lines 445-467), which is being edited concurrently in another branch.

## Test plan
- [x] `cd web && npm run test` — 41 files / 214 tests pass, including 2 new tests covering double-click and held-key-repeat guards
- [x] `cd web && npm run build` — succeeds

### DIFF
--- a/web/app/benchmark/page.test.tsx
+++ b/web/app/benchmark/page.test.tsx
@@ -226,4 +226,75 @@ describe("Benchmark page", () => {
     expect(screen.getByText("COMPILED PROMPT")).toBeTruthy();
     expect(apiJsonMock).toHaveBeenCalledTimes(2);
   });
+
+  it("ignores a second click while a benchmark run is already in flight", async () => {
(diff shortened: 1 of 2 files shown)


madara88645:

PR #987, opened 2026-07-07
fix(optimizer): local-heuristics retry no longer re-fires the failed cloud call

madara88645:
## Summary
- The Optimizer's error-banner "Switch to Local Heuristics (Offline)" button called `setProvider("local")` and then `setTimeout(() => handleOptimize(), 50)`, but `handleOptimize` closed over the render's `provider` state from before the state update, so the retry silently re-issued the exact cloud request that had just failed instead of running the offline fallback.
- `handleOptimize` now takes optional `(overrideProvider, overrideModel)` args (falling back to current state when omitted, so the main "Optimize & estimate cost" buttons and the Ctrl/Cmd+Enter shortcut keep working unchanged), and the recovery button calls `handleOptimize("local", "offline")` directly instead of relying on the setState + setTimeout race.
- Non-local optimize failures now route through `describeRequestError` (from `web/config.ts`) instead of surfacing a raw `error.message`, matching the pattern used elsewhere in the app (agent-packs, skills-generator, etc.).

## Test plan
- [x] `cd web && npm run test` — 41 files / 213 tests pass, including a new regression test asserting the recovery button runs the offline heuristic path (not another cloud call) and the cloud-error banner clears.
- [x] `cd web && npm run build` — Next.js production build succeeds.

### DIFF
--- a/web/app/__tests__/optimizer-page.test.tsx
+++ b/web/app/__tests__/optimizer-page.test.tsx
@@ -3,9 +3,18 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import OptimizerPage from "../optimizer/page";
 import { apiJson } from "@/config";
+import { showError } from "../lib/showError";
 
-vi.mock("@/config", () => ({
-  apiJson: vi.fn(),
+vi.mock("@/config", async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


madara88645:

PR #988, opened 2026-07-07
fix(web): stop clipboard-copy buttons claiming success on failure

madara88645:
## Summary
- `navigator.clipboard.writeText` rejects on unfocused tabs and non-HTTPS/insecure contexts, but several "Copy" buttons showed a success toast/checkmark unconditionally with no `.catch`.
- Add `web/app/lib/copyToClipboard.ts`, a shared helper that awaits the write, shows a success toast on success or `"Copy failed — select the text manually"` on failure, and resolves to `true`/`false` so callers only flip the "Copied!" checkmark state on an actual success.
- Route every clipboard-copy call site through it: `CopyButton`, `pr-safety`, `agent-generator`, `skills-generator`, `agent-packs`, `optimizer`, and the clipboard-copy paths in both `agent-generator`/`skills-generator` `ExportPanel.tsx` components (their unrelated error-catch/retry-button regions were left untouched).
- Existing `sr-only` / `aria-live` "Copied to clipboard" announcements are preserved — they still fire on real success.

## Test plan
- [x] `cd web && npm run test` — 42 files / 216 tests pass (added `web/app/lib/copyToClipboard.test.ts` covering success, rejection, and missing-`navigator.clipboard` cases)

### DIFF
--- a/web/app/agent-generator/components/ExportPanel.tsx
+++ b/web/app/agent-generator/components/ExportPanel.tsx
@@ -1,12 +1,12 @@
 "use client";
 
-import { toast } from "sonner";
 import { ArrowRight, Download, FolderArchive } from "lucide-react";
 
 import { useEffect, useId, useMemo, useRef, useState } from "react";
 import { useRouter } from "next/navigation";
 import { apiFetch } from "@/config";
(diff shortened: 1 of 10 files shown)


madara88645:

PR #990, opened 2026-07-07
fix(web): show persistent inline error with retry in QualityCoach

madara88645:
## Summary
- QualityCoach's `/validate` failures only fired a transient toast (`showError`), then left the panel on its pristine empty state — during a backend outage, the "Run Analysis" button looked like it silently did nothing.
- Add an `error` state to `QualityCoach`, set via `describeRequestError` (through the existing `GeneratorErrorState` component) on failure and cleared at the start of every new run.
- On failure, render an inline error card with a "Retry analysis" button, reusing `GeneratorErrorState` for visual consistency with `skills-generator`, `agent-generator`, and `pr-safety`.
- Fix the declared-but-ignored `onUpdatePrompt` prop: it was dead wiring left over from a previously removed Autofix feature (see `412c786a`, "Remove Autofix") — the component never called it. Removed it from `QualityCoachProps` and from the only call site (`web/app/page.tsx`), since there's no live feature that needs it and restoring one would be out of scope here.

## Test plan
- [x] `cd web && npm run test` — 42 files / 214 tests passed, including a new `app/components/__tests__/QualityCoach.test.tsx` covering the inline error card, the retry button (success after retry), and that a new run clears the previous error.
- [x] `cd web && npm run build` — compiles and typechecks successfully.

### DIFF
--- a/web/app/components/QualityCoach.tsx
+++ b/web/app/components/QualityCoach.tsx
@@ -3,6 +3,7 @@
 import { useState } from "react";
 import { apiJson } from "@/config";
 import { showError } from "../lib/showError";
+import GeneratorErrorState from "./GeneratorErrorState";
 
 type ValidationResponse = {
     score: number;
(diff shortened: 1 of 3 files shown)


madara88645:

PR #1062, opened 2026-07-12
🛡️ Sentinel: [MEDIUM] Fix timing attack vulnerability in API key validation

madara88645:
🚨 Severity: MEDIUM
💡 Vulnerability: Timing attack leaking admin key length
🎯 Impact: An attacker could guess the length of the admin API key based on the response time when a wrong length key is provided.
🔧 Fix: Changed the dummy comparison to use the length of the provided key instead of the admin key.
✅ Verification: Code review to ensure constant time comparison does not leak the admin key length.

---
*PR created automatically by Jules for task [14711953387638969408](https://jules.google.com/task/14711953387638969408) started by @madara88645*


madara88645:
Closing: this trades a theoretical leak for a real, remotely-triggerable 500.

**The change breaks non-ASCII API keys.** `secrets.compare_digest` raises `TypeError` on non-ASCII `str`. The old dummy compared the server's own key (`compare_digest(admin_key, admin_key)`) which is always ASCII and never raises. This PR switches it to the **caller-supplied** string (`compare_digest(normalized_key, normalized_key)`), so the operand is no longer under our control. `_matches_admin_api_key` is called at `api/auth.py:256` with no try/except, so the error escapes as a 500.

Measured both sides with a raw latin-1 `X-API-Key` header on a deployment with `ADMIN_API_KEY` set:

| request | main | this PR |
|---|---|---|
| ASCII wrong key | 403 | 403 |
| **non-ASCII key** | **403** | **500** |

main only hits that path when a non-ASCII key happens to be *exactly* `len(admin_key)`, which requires already knowing the length. This PR widens it to every non-ASCII key — a user pasting a key with an accented character gets a 500 instead of a 403.

**The leak being fixed isn't measurable.** `compare_digest` timing by length, measured locally:

```
 16 chars ->  17.6 ns/call
 32 chars ->  25.1 ns/call  (+7.5 ns)
 64 chars ->  32.3 ns/call  (+7.2 ns)
```

The admin-key length signal is ~10-40 ns against ~1-10 ms of network jitter, behind a rate limiter — roughly one part in a million. The reasoning in the PR body is sound (timing now depends on the caller's own length rather than the secret's), but MEDIUM overstates it; this is INFO at most, and not worth a 500.

PR #1168, opened 2026-07-30
🎨 Palette: Add explanation to disabled download button

madara88645:
💡 What: Added dynamic title to download button when disabled to explain why it is disabled.
🎯 Why: It wasn't obvious to users why the download button was disabled when there was no filename yet.
📸 Before/After: Visual change in tooltip.
♿ Accessibility: Improves accessibility by explaining the disabled state.

---
*PR created automatically by Jules for task [8532387006087769232](https://jules.google.com/task/8532387006087769232) started by @madara88645*


madara88645:
Closing. This is the recurring "explain the disabled button" tooltip change, reopened across several PRs.

The underlying point is reasonable — a disabled control should say why it is disabled — but it keeps arriving as an unsolicited standalone PR rather than as part of the export flow work. If it is worth doing, it should be an issue first, so the copy and the disabled-state logic get decided once instead of re-litigated per PR.

PR #1192, opened 2026-08-04
test: de-flake agent generation hard-timeout test on Windows CI

madara88645:
## Problem

`tests/test_agent_generator.py::test_worker_client_generate_agent_timeout_returns_quickly` is flaky on the `windows-latest` legs of the Full Test matrix.

Observed on `main` at ccf87b0 ([run 30915687229](https://github.com/madara88645/Compiler/actions/runs/30915687229), job *Full Test (windows-latest, 3.11)*):

```
>       assert elapsed < 0.15
E       assert 0.18378310000002784 < 0.15
```

Re-running the same job with no code change passed on all three Windows legs, so this is a flake, not a regression.

The test's intent is right — it proves the hard timeout short-circuits instead of waiting for the full call. The margin is the problem: a 0.2s stub sleep against a 0.15s budget leaves only ~0.14s of scheduling slack, and a loaded Windows runner exceeds that.

## Fix

Widen the gap rather than the tolerance:

- The `_call_api` stub now blocks on a `threading.Event` instead of `time.sleep(0.2)`, with a 10s safety net so a future bug here can never hang the suite.
- The assertion allows `elapsed < 1.0`.

The hard timeout still has to interrupt the call for the test to pass, but there is now ~1s of slack instead of a tenth. Bumping `0.15` to `0.2` was deliberately avoided — that equals the old stub's sleep and would make the assertion nearly meaningless.

The Event is released in a `finally` block so the worker thread does not linger for the full 10s after the assertion completes.

## Verification

- `python -m pytest tests/test_agent_generator.py -q` → 16 passed
- `uvx ruff@0.1.14 format --check` / `check` → clean
- Mutation check: patching `future.result(timeout=timeout_seconds)` to `future.result()` in `app/llm_engine/client.py` makes the test fail with `DID NOT RAISE` after 10.1s, confirming it still catches a genuinely broken timeout.

Note: PR checks do **not** include the Windows matrix legs (those run only on `main`), so a green PR here does not by itself prove the flake is gone — the margin argument and the post-merge `main` run are the real evidence.

### DIFF
--- a/tests/test_agent_generator.py
+++ b/tests/test_agent_generator.py
@@ -1,4 +1,5 @@
 import pytest
+import threading
 import time
 from unittest.mock import MagicMock, patch
 from app.llm_engine.client import WorkerClient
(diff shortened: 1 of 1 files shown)


madara88645:
Superseded by #1191, which landed the same fix on `main` (159debc6) while this PR was being prepared — stub sleep widened to 1.0s and the assertion relaxed to `elapsed < 0.5`.

Verified: the post-merge `main` CI run (30918555885) is green on all three `windows-latest` legs, which is the leg the flake was observed on.

This PR's variant (a `threading.Event` instead of `time.sleep`, ~1.0s of slack, worker thread released immediately after the assertion) is marginally more robust but not worth stacking a conflict resolution on top of a fix that is already proven in CI. Closing as redundant.

PR #1197, opened 2026-08-05
⚡ Bolt: Combine regex patterns for faster intent detection

madara88645:
💡 What: Combined multiple regex patterns in `_FORMAT_INTENT_PATTERNS` into a single compiled regex using `@functools.lru_cache` and replaced the `any()` generator expression with a single `search()` call.
🎯 Why: Iterating over multiple compiled regexes with a generator expression is significantly slower than pushing the OR iteration down to the C-based regex engine using a single combined pattern.
📊 Impact: Expected to reduce execution time for `is_output_formatting_intent` by ~2x for matches and ~3x for non-matches.
🔬 Measurement: Verify tests pass with `pytest tests/test_skill_output_fidelity.py` and measure execution time of `is_output_formatting_intent` against the previous `any()` implementation.

---
*PR created automatically by Jules for task [2207701259547453577](https://jules.google.com/task/2207701259547453577) started by @madara88645*

### DIFF
--- a/app/llm_engine/skill_output_fidelity.py
+++ b/app/llm_engine/skill_output_fidelity.py
@@ -10,6 +10,7 @@ preserved content so tests can lock the contract without calling an LLM.
 
 from __future__ import annotations
 
+import functools
 import re
 import textwrap
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


madara88645:

PR #1199, opened 2026-08-06
⚡ Bolt: [regex performance optimization]

madara88645:
💡 What: Combined multiple regex patterns into a single compiled pattern using the regex OR operator.\n🎯 Why: Checking a string against multiple regular expressions iteratively using an \`any()\` generator expression introduces unnecessary overhead from generator setup/teardown and redundant regex compilation/evaluation.\n📊 Impact: Pushes the search iteration directly to the C-based regex engine, which is much faster.\n🔬 Measurement: Run the backend test suite.

---
*PR created automatically by Jules for task [5075574963980897398](https://jules.google.com/task/5075574963980897398) started by @madara88645*


madara88645:

PR #1239, opened 2026-08-15
test: add unit coverage for export-adapter private helper functions

madara88645:
## Summary

Adds unit test coverage for pure, previously-untested private helper functions in the export-adapter layer, found during a test-coverage audit. Several of these directly affect generated-code correctness (Python identifier safety, triple-quote escaping, markdown-fence stripping) so leaving them untested carried real regression risk. New test files only — no source, config, or CI changes.

## Changes

- `tests/test_skill_adapter_private_helpers.py` (58 tests) — covers `app/adapters/skill_adapter.py`: `_needs_keyword_only_params`, `_yaml_safe`, `_truncate_at_sentence`, `_capitalize_words`/`_to_kebab`/`_to_title`, `_unwrap_to_base`.
- `tests/test_claude_code_private_helpers.py` (33 tests) — covers `app/adapters/claude_code.py`: `_to_python_identifier`, `_strip_markdown_fences`, `_escape_triple_quotes`; `app/adapters/agent_ir.py`: `_map_sections`; `app/adapters/agent_packs.py`: `_preview_order` (`_classify_kind` in the same module was already covered by the existing `TestClassifyKind` suite, so it wasn't duplicated).

One documented, non-blocking behavioral note: `_strip_markdown_fences("```\n```")` (a fenced block with literally empty content) returns `` ``` `` instead of an empty string, because the leading-fence regex consumes the shared newline before the trailing-fence regex can match. This is asserted as verified real behavior (`test_empty_fenced_block_leaves_closing_backticks`), not silently treated as a bug — it only affects a degenerate empty-fence input and doesn't block this PR, but may be worth a follow-up if fully-empty fenced LLM output needs to be handled.

## Testing

- New suite: `python -m pytest tests/test_skill_adapter_private_helpers.py tests/test_claude_code_private_helpers.py -v` → 91 passed
- Full suite: `python -m pytest tests/ -q` → 2640 passed, 7 skipped (baseline before this change: 2549 passed, 7 skipped — exactly +91, zero regressions)

---
_Generated by [Claude Code](https://claude.ai/code/session_01TWn8B6e1jgGuSpdhqVgpF4)_

### DIFF
--- a/tests/test_claude_code_private_helpers.py
+++ b/tests/test_claude_code_private_helpers.py
@@ -1,0 +1,227 @@
+"""
+Coverage for small, pure, string-shaping helpers that were not directly
+exercised by any existing test:
+
+  * app/adapters/claude_code.py — _to_python_identifier, _strip_markdown_fences,
+    _escape_triple_quotes
+  * app/adapters/agent_ir.py    — _map_sections
+  * app/adapters/agent_packs.py — _preview_order
+
+Note: app/adapters/agent_packs.py's `_classify_kind` is intentionally NOT
+re-tested here — it already has a dedicated, exhaustive test class
+(`TestClassifyKind`) in tests/test_agent_packs_private_helpers.py.
+"""
+
+from __future__ import annotations
+
+from app.adapters.agent_ir import _map_sections
+from app.adapters.agent_packs import _preview_order
+from app.adapters.claude_code import (
+    _escape_triple_quotes,
+    _strip_markdown_fences,
+    _to_python_identifier,
+)
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


madara88645:
Closing as part of a one-off coverage-backlog cleanup, not as a review verdict on this change.

Three separate automations (Cursor, agent sessions, and the `tests/add-*` job) were independently generating coverage for the same modules. Because each wrote to a differently-named test file they never collided, so the queue silently grew to 35 overlapping drafts across three repositories — none of which ever left draft state.

Remaining coverage gaps will be reopened as a single consolidated PR per repository. The branch is left in place if any of this needs to be recovered.

